PR #3, opened 2019-08-19
Fix default path for docker image and update README

### LINKED ISSUE #2: Adam runtime file location doesn't match documentation
The Adam server uses /run/adam for runtime files, rather than /adam/run/adam as described in the documentation. Elsewhere in the documentation a relative path ./run/adam without specifying what the PWD is inside the docker container, leading one to assume it is /adam rather than / which is currently is.

deitch:
Fixes #2 

Please take a look @rvs ; cc @mhall119


deitch:
@rvs the only checks running are DCO, but there is a whole set of circleCI checks in the config. Can you enable?

PR #4, opened 2020-02-17
Change config handler from GET to POST

cshari-zededa:
- also fill in ConfigResponse as the response to POST

Signed-off-by: Hariharasubramanian C S <cshari@zededa.com>

### DIFF
--- a/pkg/server/apiHandler.go
+++ b/pkg/server/apiHandler.go
@@ -4,15 +4,19 @@
 package server
 
 import (
+	"crypto/sha256"
 	"crypto/x509"
 	"encoding/base64"
+	"encoding/json"
 	"encoding/pem"
+	"fmt"
 	"io/ioutil"
 	"log"
 	"net/http"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/lf-edge/eve/api/go/info"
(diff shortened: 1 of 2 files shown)


deitch:
I assume this is to align with [eve#641](https://github.com/lf-edge/eve/pull/641) for the api and [eve#550](https://github.com/lf-edge/eve/pull/550) for the client side implementation?

Given the above, this makes sense. However, should we be keeping the deprecated method around for a bit?

pkg/server/apiHandler.go
```diff
@@ -113,16 +117,25 @@ func (h *apiHandler) config(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
-	out, err := proto.Marshal(config)
+	configResp := &config.ConfigResponse{}
+	blob, err := json.Marshal(msg)
+	if err != nil {
+		log.Printf("error marshaling config to json: %v", err)
+	}
+	checksum := fmt.Sprintf("%X", sha256.Sum256(blob))
+	configResp.Config = msg
```
eriknordmark: This should only be included if the checksum is different than what was in the configRequest.

cshari-zededa:
closing this, as PR#5 is a superset of this

PR #9, opened 2020-04-29
Introducing EVE+Adam embedded mode of running Adam

rvs:
This PR introduced an extra mode of running Adam in an environment where it is running on the same VM with all the rest of the EVE services. This is useful in variety of different scenarios to make EVE+Adam pair self-contained and available to accept push configuration instead of pull one. For example, the way I'm using this right now is  by building EVE rootfs image with the following extra section in the `services:` block:

```
  - name: adam
     image: lfedge/adam:latest
     binds:
        - /var/config:/config
        - /var/persist:/persist
     command: ["/bin/eve-embedded.sh"]
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,9 +22,10 @@ ARG GOARCH=amd64
 
 RUN go build -o /adam/bin/adam main.go
 
-FROM scratch
+FROM alpine:3.11
 
-COPY --from=build /adam/bin/adam /adam/bin/adam
+COPY scripts /bin
+COPY samples /adam
+COPY --from=build /adam/bin/adam /bin/
 WORKDIR /adam
-ENTRYPOINT ["/adam/bin/adam"]
-
+ENTRYPOINT ["/bin/adam"]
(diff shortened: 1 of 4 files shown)


deitch:
Let me see if I get this. The job here is to simplify, such that if adam and eve are running on the same host - and therefore have access to common filesystems, notably adam can access eve's `/config`,  then you can use a start script in the container to wrap the `adam` command and put all of its certs and config in the right place (`/config`) so that eve is configured and ready to talk to that adam?

.gitignore
```diff
@@ -1,3 +1,3 @@
 run/
-bin/
+bin/adam
```
deitch, line 2: You can build for alternate OSes, so let's leave `bin/` as the entire directory that is in `.gitignore`. No reason we cannot put `eve-embedded.sh` in the root directory here.

If you really want it in `bin/`, then let's make the output of the builds in `dist/` and save `eve-embedded.sh` in bin/`, but make sure to change the `Makefile` and the README.doc

rvs: I am not a big fan of lose files in the root folder, so I added it to scripts. Hope this is OK @deitch

rvs:
Took care of your concerns @deitch -- please take a look now

.gitignore
deitch: Sure, that is just as fine.

deitch:
Looks good, thanks for updating @rvs .

We might be able to simplify it further by building some of this functionality directly into adam. If I have a chance, I will look at a later date.

PR #13, opened 2020-05-01
Cleaning up go.mod (go mod tidy) and latching onto the latest EVE API

rvs:
This should be pretty self-explanatory


rvs:
Hey @deitch -- I'm taking a liberty to self-approve this -- since I'd live to cut an EVE/Adam/Eden release trio that actually is synced up on the API usage. We can fix it later if this breaks anything.

Issue #16, opened 2020-05-04
Need to implement V2 API in Adam


rvs:
EVE has transitioned to the V2 API by default. Adam knows only about V1 still. This means we have to force all instances of EVE that talk to Adam to have Force-API-V1 in the config folder -- otherwise EVE won't be able to talk to Adam at all.

There's work in EVE going on to publish V2 API spec updates -- once that is done we'll use the spec to update Adam to V2 API.

deitch:
@rvs not sure I understand this. Is the below correct?

-  eve can talk v1 and v2, but uses v2 by default
- if eve receives a particular signal on onboarding, it falls back to v1
- what is that signal?
- can other controllers (zedcloud) talk both?
-  can and should we get adam to talk both?
- since the api is driven by generated protobuf files, how do they coexist?

rvs:
Hey @deitch -- here's the deal:

> eve can talk v1 and v2, but uses v2 by default

correct -- in addition to that there's an agreement that v1 support goes away soon (once Adam gets a chance to catch up basically)

> if eve receives a particular signal on onboarding, it falls back to v1

nope -- onboarding itself is part of the API hence we can't use that as a signal -- the only we can give EVE to fallback onto V1 is that file in /config

> can other controllers (zedcloud) talk both?

Zedcloud is in a process of transitioning to V2 only

> can and should we get adam to talk both?

there's not really much value, since EVE is going to drop V1 the moment Adam can support V2

> since the api is driven by generated protobuf files, how do they coexist?

The API are on different endpoints

@eriknordmark and Naiming are documenting all of this in this current sprint

deitch:
> onboarding itself is part of the API hence we can't use that as a signal -- the only we can give EVE to fallback onto V1 is that file in /config

That aligns with what you did in #17 ; makes sense.

> Zedcloud is in a process of transitioning to V2 only
> The API are on different endpoints

So we are just moving to v2 and retiring v1. Fine with me.

> @eriknordmark and Naiming are documenting all of this in this current sprint

So the position is that we hold until that is documented, then implement it here?

PR #17, opened 2020-05-05
Forcing EVE to use V1 API while #16 is getting fixed

rvs:
This also simplifies the handshake between EVE and Adam (especially for the embedded case)

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -79,6 +79,21 @@ var serverCmd = &cobra.Command{
 		if err != nil {
 			log.Fatalf("error writing hosts file: %v", err)
 		}
+
+		rootCert, err := ioutil.ReadFile(serverCert)
+		if err != nil {
+			log.Fatalf("error reading %s file: %v", serverCert, err)
+		}
+		err = ioutil.WriteFile(path.Join(configDir, "root-certificate.pem"), rootCert, 0644)
+		if err != nil {
+			log.Fatalf("error writing root-certificate.pem file: %v", err)
+		}
+
+		// FIXME: this is going away as part of fixing https://github.com/lf-edge/adam/issues/16
+		err = ioutil.WriteFile(path.Join(configDir, "Force-API-V1"), []byte{}, 0644)
+		if err != nil {
+			log.Fatalf("error writing Force-API-V1 file: %v", err)
+		}
 		log.Printf("EVE-compatible configuration directory output to %s", configDir)
 
 		s := &server.Server{
--- a/scripts/eve-embedded.sh
+++ b/scripts/eve-embedded.sh
@@ -41,10 +41,8 @@ bootstrap() {
 
 # if this is the first run on this /persist -- generate everything
 if [ ! -d $DB ]; then
-   adam generate --db-url $DB --server --hosts 127.0.0.1,localhost --cn localhost.localdomain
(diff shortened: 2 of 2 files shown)


scripts/eve-embedded.sh
```diff
@@ -41,10 +41,8 @@ bootstrap() {
 
 # if this is the first run on this /persist -- generate everything
 if [ ! -d $DB ]; then
-   adam generate --db-url $DB --server --hosts 127.0.0.1,localhost --cn localhost.localdomain
-   cp $DB/server.pem /config/root-certificate.pem
-   echo $SERVER > /config/server
```
deitch, line 46: Isn't removing this going to break it? The default is not `localhost:6000`

cmd/server.go
```diff
@@ -79,6 +79,21 @@ var serverCmd = &cobra.Command{
...
+		if err != nil {
+			log.Fatalf("error reading %s file: %v", serverCert, err)
+		}
+		err = ioutil.WriteFile(path.Join(configDir, "root-certificate.pem"), rootCert, 0644)
+		if err != nil {
+			log.Fatalf("error writing root-certificate.pem file: %v", err)
+		}
+
+		// FIXME: this is going away as part of fixing https://github.com/lf-edge/adam/issues/16
+		err = ioutil.WriteFile(path.Join(configDir, "Force-API-V1"), []byte{}, 0644)
```
deitch, line 93: Should these not go into `eve-embedded.sh` rather than here? This `/config/Force-API-V1` file is only in the embedded context, and only makes sense to eve itself. 

I don't object to adding some capabilities to adam itself, so that we can eventually get rid of `eve-embedded.sh` (I was thinking about it, actually), but let's not put them in in an unstructured way. This whole thing can be done in a single `cp` command and a single `echo` command in `eve-embedded.sh`

rvs: No it actually has to go everywhere -- not just into embedded. Otherwise there's absolutely no way for us to signal to the user that this needs to be somehow handed over to EVE via its config.

In fact, that the whole point of Adam generating an EVE config -- it needs to have a way to signal of what options need to be there for it to actually function.

Hence I think it actually belongs here -- the only other alternative @deitch is to completely drop this EVE configuration generation -- 'cuz it really has to be complete or it needs to be dropped.

The thing we don't want it to be is misleading -- that's the worst option of all.

scripts/eve-embedded.sh
rvs: It is very much localhost:6000 in the embedded case

rvs:
> I get what you are trying to do, but I don't think this goes in adam itself, or at least not yet. Let's keep all of the "embedded" 

Well, but the problem is -- its not just embedded. Eden folks ran into issue here as well. Like I said -- we need to make sure that Adam either generates all config knobs that need to be given to EVE for it to actually function.

OR 

We need to completely drop EVE config generation from Adam.

For now -- I'd prefer option #1

scripts/eve-embedded.sh
deitch: Where is that set?

cmd/server.go
deitch: Adam already was writing out an eve-compatible config to a local directory when run in `adam server` mode. Right. In that case, this makes sense.

PR #21, opened 2020-05-07
logfiles and info with maxima

deitch:
Provides for rotating log files and log in memory, configurable maximum size, and rotation. Also does a lot of under the covers clean up in those spaces.


deitch:
@rvs I want to try this with your embedded adam, which I haven't had the chance to do yet. So don't merge it until I do.

Was your embedded merged into eve yet? Or do I still need to build from your branch?

rvs:
Hey @deitch -- EVE side has been merged so you should be able to do:
```
make HV=adam-kvm live
make ACCEL=y PFLASH=y run
[wait for at least a minute]
make run-proxy
```

deitch:
This passes all of my tests. The log rotation size is not working correctly - it should be much larger - but it is still vastly improved over the current one (one file per message), so we can merge this in and fine-tune it later. Back to you @rvs

PR #23, opened 2020-05-10
Make better onboarding experience from newbies. Updated Readme

mydatascience:
As Adam is not in any package management system like apt or yum most often the user will build it first and then run. 

Also someone may not read carefully the TLS  part and by default the _adam generate server_ command doesn't generate the certificate files in the correct destination. (--server option is not default) So updated the command to the most likely used.


deitch:
Why did you close it?

PR #24, opened 2020-05-10
Make better onboarding experience for newbies. Updated Readme 

mydatascience:
As Adam is not in any package management system like apt or yum most often the user will build it first and then run.

Also someone may not read carefully the TLS part and by default the adam generate server command doesn't generate the certificate files in the correct destination. (--server option is not default) So updated the command to the most likely used.


rvs:
I OK with this. @deitch ?

deitch:
So this primarily moves the building part from after "running" to before "running"? I don't mind terribly, but I think a better experience would be having a release cut. FYI, every successful merge to master does create a docker image, so should "building it" be first?

mydatascience:
If we focus on the developers they would likely do `git clone` and build it first. 

If you think it's better to have run first -  then let's move TLS part before run part. It doesn't work without TLS certs. Then if I don't have ./bin/adam locally then I will not be able to have a docker image work.   

Also
If we focus on users that will deploy it from the Dockerhub then probably we should provide the Overview.

deitch:
> If you think it's better to have run first

Definitely. We don't want people to start by seeing building, and then thinking, "I need to be a developer to run this? git? huh??"

Take a look at #29 and the doc changes there, which were partially done to address this. Then we can iterate on that.

mydatascience:
Your #29 is a good solution. Let's go from there.

rvs:
I just merged #29 -- should this be closed or reworked @mydatascience ?

rvs:
Ping @mydatascience -- please let us know how do you want to proceed?

mydatascience:
I'm ok for now. Let's close. I may work with the newer version

PR #30, opened 2020-05-14
Introducing Redis backend

rvs:
This introduces Redis DB as a backend for Adam. Think about this as a way to externalize internal Adam's state in a such a way that it is very accessible through the network (by things like Eden even scripts). In addition to that we're using Redis streams as poor man's Kafka to manage our append logs (logs, info and messages). This improves the usability quite a bit sine we get nifty things like offsets, blocking/non-blocking reads and consumer groups.

This is not quite ready to be merged, but definitely ready to be reviewed @deitch @giggsoff


deitch:
I didn't merge it, since you said it isn't ready. I will hold on restructuring the `driver/` dir until this is.

rvs:
I am running a few tests to determine the most reliable persistence model that at the same time wouldn't get too much in the way of performance/resource consuption.

One place where I definitely need your collective help is figuring out how to integrate this in the Docker. On one hand, we can take a hard stance and say that integration has to happen in the "cloud native way" and Adam container has no business shipping Redis (and you'll have to do the integration via docker-compose or something)

On the other hand -- the way I'm envisioning using Redis almost makes it an integral (embedded if you will) part of Adam. IOW, it Redis was written in Go -- I'd just import it statically into Adam as opposed to running as a side-car. From that standpoint may be should have a script entry point into Adam container that always launches Redis whenever the 'server' CLI option is give to the container.

@deitch @sadov @giggsoff @mydatascience what do you guys think?

deitch:
The cloud native way is to have adam do just adam, and redis is external. Run it if you want. Besides, what happens when someone what’s MySQL? Will we then add that? And oracle? And mongo? I don’t mind embedding the drivers, but the actual databases? Not likely. 

The only thing I would think of embedding is sqlite because it is embedded natively. Even then, I’d be concerned because it’s natively C, which materially complicates building adam. Unless someone rewrote it in go?

mydatascience:
I'd say - make it an option and require an external redis within a new docker file/ redis container on EVE for Embedded Adam.
Agree with @deitch we embed connectors. Also it's better to save option of logging without redis.

mydatascience:
https://wiki.lfedge.org/display/LE/Logging 
I suggest to reference this document for the implementation.

PR #34, opened 2020-05-29
Update README.md

dcmartin:
Wrong pathname for `config/` directory; should be `run/config` and not `run/adam/config`


deitch:
Nice catch, thank you for fixing.

Please sign your commit with `git commit -s` and re-push, and we can merge it in.

rvs:
Ping @dcmartin -- any chance you can sign your commit so we can pull it in?

deitch:
This is old and untouched.

PR #41, opened 2020-09-08
add support for tracking all requests

deitch:
Tracks the IP, X-Forwarded-For header, timestamp and path for all `/api/*` requests. Stores them the same way as logs, info, metrics, and uses the same method to retrieve them.

Also adds admin endpoint and cmd flag to retrieve them, similar to logs, info, metrics.

All of the above should have an ability to get the first or last X, but that is a separate question.

The results are json, and can be passed through jq, e.g.:

```json
{
  "ip": "172.31.15.153:46936",
  "method": "POST",
  "url": "/api/v1/edgedevice/config"
}
{
  "ip": "172.31.15.153:46940",
  "method": "POST",
  "url": "/api/v1/edgedevice/config"
}
{
  "ip": "172.31.15.153:46942",
  "method": "POST",
  "url": "/api/v1/edgedevice/config"
}
{
  "ip": "172.31.15.153:46944",
  "method": "POST",
  "url": "/api/v1/edgedevice/metrics"
}
```


rvs (approved):
Very nice! Merging.

PR #42, opened 2020-09-10
fix bigdata driver and unify implementation

deitch:
Does two things:

* unifies the Redis backend driver's usage of `BigData struct` for (logs, info, metrics, requests) to be aligned with the others (file, memory). In the process, eliminates a lot of duplication
* fixes where the `BigData struct` had a `Read()` call that never was used. Cleans it up to be a `Reader()` and get a standard `io.Reader()`

Once this is in, will enable it to get reader for, say, "last 5 entries".

### DIFF
--- a/pkg/driver/common/common.go
+++ b/pkg/driver/common/common.go
@@ -8,6 +8,7 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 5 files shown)


rvs (approved):
LGTM. Please pull it @deitch

Issue #47, opened 2020-11-23
Link to "API-compliant Controller" isn't clickable


mperov:
## Description
There is link to API-compliant Controller in README.md, but I can't access to [API-compliant Controller](https://github.com/lf-edge/eve/blob/master/api/API.md)
I get **404 Error - Page not found**!

mperov:
Link to **the official API** isn't working too.

rvs:
Seems like typos. They should be pointing to https://github.com/lf-edge/eve/blob/master/api/README.md

Do you want to submit a Pull Request @mperov ?

mperov:
@rvs yes, I do, so I submitted pull request from my fork of adam, please, consider it.

deitch:
Closed via #49

PR #48, opened 2020-11-24
some typos are fixed in README.md

mperov:
I'd like to fix typos are described in #47


mperov:
ok, I'll create new pull request and be careful.

PR #54, opened 2020-12-06
check URL endpoint in file manager; put test artifacts in tmpdir

deitch:
Two things:

* checks the URL endpoint in file manager. Before, it was just taking the string as is, which creates some really strange behaviours, such as creating a directory named `redis:`. Now, it checks the URL. It was supposed to do this originally.
* For the one place in tests where it actually creates a directory, do it in a tempdir with an automatic cleanup.

### DIFF
--- a/pkg/driver/device_managers_test.go
+++ b/pkg/driver/device_managers_test.go
@@ -1,6 +1,9 @@
 package driver_test
 
 import (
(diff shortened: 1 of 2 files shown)


rvs (approved):
LGTM

PR #56, opened 2020-12-07
switch from circleci to github actions

deitch:
circleci wasn't getting executed, but we needed to move here anyways.

Creates build (CI), publish (merge to master) and release (tags) workflows. Publishes to docker hub.

Needs a few things. @rvs can you review this? I assume it will take a few runs to get this right (it always does with changes to the pipeline itself).

We also need the hub credentials in `secrets.DOCKERHUB_USERNAME` and `secrets.DOCKERHUB_TOKEN`. I don't know if you have that @rvs or @aw-was-here has it?


deitch:
Oh, yes, we also need someone with admin rights to enable it and make it required.

rvs (approved):
LGTM

PR #57, opened 2020-12-07
Fixing a tiny typo of everything being called build

rvs:
Signed-off-by: Roman Shaposhnik <rvs@zededa.com>


deitch (approved):
Nice catch.

deitch:
CI is failing because of fmt, which is exactly what this PR is supposed to catch. Going to merge in and we can fix that issue separately.

PR #62, opened 2020-12-08
report gofmt output

deitch:
The `fmt` check was checking if formatting is good, but not reporting the output to logs. Of course, someone always can rerun it locally - which is what we did beforehand - but it really should report the error on the logs, rather than confusing someone.


deitch:
The error in CI for this PR is because it actually is picking up formatting errors. That is fixed in #61

rvs (approved):
LGTM

PR #74, opened 2021-02-26
Support swagger ui

mydatascience:
This is some work to make Adam UI nicer


Dockerfile
```diff
@@ -21,11 +21,16 @@ ARG GOOS=linux
...
+RUN go get -u github.com/go-swagger/go-swagger/cmd/swagger
+RUN GO111MODULE=off swagger generate spec -o ./swaggerui/swagger.json
+
 
 FROM alpine:3.11
 
 COPY scripts /bin
 COPY samples /adam
+ADD swaggerui ./swaggerui/
+COPY swaggerui /bin/
```
deitch, line 33: Why are we putting the same `swaggerui` dir in 2 separate locations in the container?

Dockerfile
```diff
@@ -21,11 +21,16 @@ ARG GOOS=linux
 # ARG GOARCH=amd64
 
 RUN go build -o /adam/bin/adam main.go
+RUN go get -u github.com/go-swagger/go-swagger/cmd/swagger
+RUN GO111MODULE=off swagger generate spec -o ./swaggerui/swagger.json
```
deitch, line 25: It looks like this generates `swaggerui/swagger.json`, but then this is never copied from the build image to the final image.

swaggerui/swagger.json
```diff
@@ -0,0 +1,58 @@
+{
```
deitch, line 1: Is this file auto-generated? It has it generated inside the build container, but also is here.

If it is auto-generated, why do we need it? Unless we are assuming that we want it checked in for running locally?

deitch (changes requested):
There were a few things that I didn't understand why they were there, or duplicated. If they were errors (probably not), please fix them. If they were intentional (probably), please add comments to clarify.

The other request I have is that we add documentation describing how we use swagger here. We don't need to copy all of  swagger docs, just some basics such as:

* do we have swagger.json locally, or just in the build container, and what is the relationship between them?
* Is it checked in, and, if so, what is the right way to regenerate it, and when _should_ it be checked in?
* what process should someone follow if they want to add more API endpoints, whether root or under existing ones, in terms of the need to ensure that they end up in swagger?

rvs:
Hey @mydatascience -- what's the latest on this?

deitch:
Bump?

Dockerfile
mydatascience: Fixed

Dockerfile
mydatascience: Yes I removed it

swaggerui/swagger.json
mydatascience: yes. removed

mydatascience:
Fixed. @deitch @rvs

deitch (approved):
LGTM.

Does this mean we will need to annotate every single call we expose? That could be more than middle painful.

mydatascience:
Yes we do. An annotation is mostly about comments 
What can be derived is derived. 
Look at the sample below: 

```

// swagger:operation GET /onboard onboard
	//
	//
	// Onboards EVE
	//
	// The EVE should connect first
	//
	// ---
	// produces:
	// - application/json
	// tags:
	// - admin
	// responses:
	//   '200':
	//     description: successful operation
	ad.HandleFunc("/onboard", admin.onboardList).Methods("GET")


// OnboardCert encoding for sending an onboard cert and serials via json
// swagger:parameters onboard
type OnboardCert struct {
	// a Cert for onboarding
	//
	// unique: true
	// in: query
	Cert []byte
	// a Serial for onboarding
	//
	// unique: true
	// in: query
	Serial string
}
```

mydatascience:
Another way is to design API json and do code generation. The other way (generate description from the code and comments) looks more appealing to me

deitch:
Both have burdens, but let's at least start with what you have. No one says we need perfect, just incremental improvements, and what you have improves it.

Is this ready to go in @mydatascience ?

mydatascience:
Yes please merge. Will have a new one soon

deitch:
Please squash these down to a single commit first.

mydatascience:
Done @deitch.

deitch (approved):

deitch:
@mydatascience the merge action failed. See [here](https://github.com/lf-edge/adam/actions/runs/697155323). Please fix and open a PR as needed.

PR #83, opened 2021-04-02
[WIP] New v2 implementation

mydatascience:
Support proto AuthContainer and v2 API

### DIFF
--- a/pkg/server/apiHandlerv2.go
+++ b/pkg/server/apiHandlerv2.go
@@ -1,0 +1,495 @@
+// Copyright (c) 2019 Zededa, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package server
+
+import (
+	"bufio"
+	"compress/gzip"
+	"crypto/sha256"
+	"crypto/x509"
+	"encoding/base64"
+	"encoding/json"
+	"encoding/pem"
+	"fmt"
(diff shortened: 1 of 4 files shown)


deitch:
Thanks for getting this moving, @mydatascience.

I have a few questions:

* we will support both v1 and v2, with v1 at `/apps/v1/edgedevice/` and v2 at `/apps/v2/edgedevice/`. Is that correct?
* the endpoints under those paths are identical, correct?
* `ApiRequest` and `ApiRequestv2` are identical, as are each of the `recordClient` implementations. Unless there is a strong reason not to do so, let's pull them out of `(h *apiHandler)`/`(h *apiHandlerV2)` and make it a single shared utility function
* `checkCertAndRecord` are identical for both `apiHandler` and `apiHandlerV2`. Unless there is a strong reason not to do so, let's pull them out and make it a single shared utility function, passing in `h.manager` as an arg
* Why do we have the two protobuf files here? Don't they already exist in github.com/pkg/eve/api/`, plus go-compiled version? We should just be importing what we need, and in go, not protobuf, if possible.

Actually, going through this in detail, I don't see a single difference between the `apiHandler` and `apiHandlerV2` implementation. What does this add? I see just some some unused protobuf files, and some endpoints that act exactly like v1.

PR #86, opened 2021-04-19
use latest on semver and snapshot on every master branch update

deitch:
As discussed in #85 

Once this is in, we will cut a release.


rvs (approved):
LGTM

PR #125, opened 2024-05-02
Fix broken links to EVE API docs

stmcginnis:
The eve-api was broken out into its own repo last year. There were a few links in the adam docs that were still pointing to their old location, leading to page not found errors. This updates the links to point to their current locations.


deitch (approved):

deitch:
Thank you for this, and good catch. As soon as CI is green, we can merge it in.

PR #126, opened 2024-06-06
when keys not found in redis, return NotFoundError

deitch:
The file and memory driver distinguish between actual errors and things that are not found by returning a `common.NotFoundError`, which the admin handler check for.

The redis driver does not; for everything that it fails to get, it just returns an error. 

The net result, e.g, is if you look up a certificate for a device that does not exist, with the memory or file driver, adam will return a 404, while the redis driver will return a 400, which is incorrect. Downstream things like eden then think it is an error, rather than something just not found yet.

This adds `NotFoundError` to places in redis where `HGet` and its variants return `redis.Nil`, which means, "I could not find the key."

### DIFF
--- a/pkg/driver/redis/device_manager_redis.go
+++ b/pkg/driver/redis/device_manager_redis.go
@@ -222,6 +222,9 @@ func (d *DeviceManager) OnboardGet(cn string) (*x509.Certificate, []string, erro
 	}
 
 	s, err := d.client.HGet(onboardSerialsHash, cn).Result()
+	if err != nil && errors.Is(err, redis.Nil) {
+		return nil, nil, &common.NotFoundError{Err: fmt.Sprintf("device not found %s", cn)}
+	}
 	if err != nil {
 		return nil, nil, fmt.Errorf("error reading onboard serials for %s: %v", cn, err)
 	}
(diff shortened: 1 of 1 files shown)


pkg/driver/redis/device_manager_redis.go
```diff
@@ -658,6 +664,9 @@ func (d *DeviceManager) GetCerts(u uuid.UUID) ([]byte, error) {
 	// hold our config
 	var b []byte
 	data, err := d.client.HGet(deviceAttestCertsHash, u.String()).Result()
+	if err != nil && errors.Is(err, redis.Nil) {
+		return nil, &common.NotFoundError{Err: fmt.Sprintf("device not found %s", u)}
```
giggsoff, line 668: Maybe `attestation certificates for the device %s not found` will better describe the error here?

pkg/driver/redis/device_manager_redis.go
```diff
@@ -699,6 +708,9 @@ func (d *DeviceManager) GetStorageKeys(u uuid.UUID) ([]byte, error) {
 	// hold our config
 	var b []byte
 	data, err := d.client.HGet(deviceStorageKeysHash, u.String()).Result()
+	if err != nil && errors.Is(err, redis.Nil) {
+		return nil, &common.NotFoundError{Err: fmt.Sprintf("device not found %s", u)}
```
giggsoff, line 712: Maybe `storage keys for the device %s not found` will better describe the error here?

giggsoff (commented):
I am not sure, but will we hit the same problem with calls to `readCert` function for example in `DeviceGet`?

pkg/driver/redis/device_manager_redis.go
deitch: Sure, I like this comment and the next. Adding them now.

deitch: Done both.

pkg/driver/redis/device_manager_redis.go
deitch: Done

giggsoff (approved):

PR #127, opened 2024-07-10
GitHub Actions use org secrets

deitch:
lfedge created a new user for all of the workflows, and saved the credentials as org secrets, specifically `RELEASE_DOCKERHUB_ACCOUNT` and `RELEASE_DOCKERHUB_TOKEN`. This PR removes the different name in the flows and uses the org standard `secrets.RELEASE_DOCKERHUB_ACCOUNT` and `secrets.RELEASE_DOCKERHUB_TOKEN`.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -25,8 +25,8 @@ jobs:
       - name: Login to DockerHub
         uses: docker/login-action@v1 
         with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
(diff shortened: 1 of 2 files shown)


yash-zededa (approved):

milan-zededa (approved):

milan-zededa:
Maybe we should use never version of the docker-login action to avoid the warning and problems in the future: https://github.com/lf-edge/adam/actions/runs/9871707714/job/27260199006#step:5:10

deitch:
Yes, we definitely should. This PR was very focused, fix that one thing. Another PR for that would be good.

PR #132, opened 2024-11-28
gha: release.yml fix docker tag

uncleDecart:
before release.yml was never triggered, so error was not visible this patch should fix it.


uncleDecart:
That'll be a bunch of merges and releases...

deitch (approved):
Welcome to the world of pipeline fixes in GHA. 😆

PR #133, opened 2024-11-28
gha: release.yml fix ghcr.io tag

uncleDecart:
Take 2 :) 

previously used variable was not available.


deitch (approved):
As long as you are sure. These github vars always are messy and I get them wrong.

PR #136, opened 2024-12-10
apiHandlerv2 : make sure register request is signed

shjala:
This change makes sure that the register request is signed with onboarding cert, that is the only purpose of onboarding cert.


uncleDecart (approved):
LGTM

PR #143, opened 2025-09-26
bump golang.org/x deps

deitch:
Supersedes #141


milan-zededa (dismissed):

deitch:
Yes, looks like it. Adam in dockerfile is based on an old eve-alpine, which has an older go 1.20. I am seeing what the most recent has, and will bump that, too, if it solves. Should know in about 2 mins.

deitch:
OK, cool, this is a different error!

milan-zededa (approved):

PR #146, opened 2026-04-17
Implement SCEP proxy endpoint

milan-zededa:
Add `POST /api/v2/edgedevice/id/{uuid}/proxy/scep` that lets an Edge Device perform SCEP (Simple Certificate Enrollment Protocol) operations through the Controller when the SCEP server is not directly reachable from the device network (e.g., during onboarding on an unauthenticated VLAN).

The Controller acts as a pure transport-level proxy: it authenticates the device via the standard `AuthContainer` mechanism, verifies that the requested SCEP profile is authorized in the device configuration, then forwards the SCEP PKI message to the configured SCEP server URL and relays the response back wrapped in an `AuthContainer`. The PKCS#7 payload is never unwrapped, decrypted, or inspected by the Controller.

Request and response bodies carry `proxy.SCEPProxyRequest` / `proxy.SCEPProxyResponse` protobuf messages (eve-api `go/proxy` package) inside the standard `AuthContainer` envelope. Both GET (`GetCACaps`, `GetCACert`, `GetNextCACert`) and POST (`PKIOperation`) SCEP operations are supported. Query parameters and HTTP headers from the device request are forwarded to the upstream SCEP server verbatim.

Bump eve-api dependency to pick up the new proxy protobuf definitions.

Security properties (per spec):
- Device private key never leaves the device (TPM-backed when available)
- All SCEP crypto (signing, decryption) is performed on the device
- Controller provides only device authentication, profile authorization, and transport proxying

Implements the protocol defined in [APIv2.md §"SCEP Proxy"](https://github.com/lf-edge/eve-api/blob/main/APIv2.md#scep-proxy).


uncleDecart (approved):
LGTM

PR #147, opened 2026-04-17
Support multiple concurrent EVE devices

milan-zededa:
# Summary

This PR makes Adam safe for use with multiple EVE devices connecting and
reporting data simultaneously. It addresses two orthogonal concerns:
- correct per-device data routing in the admin streaming interface, and
- race-free access to the in-memory state shared by concurrent device goroutines.

# Changes

## Per-device admin streaming (`pkg/server/`)

Previously, the four live-data channels (`log`, `info`, `metric`,
`requests`) were single global `chan []byte` instances shared across all
devices. Any admin SSE client watching one device would receive data from
all devices indiscriminately, and multiple admin clients would race for
the same events.

**`pkg/server/stream.go`** -- new `stream` type: a UUID-keyed,
thread-safe publish/subscribe multiplexer. `stream.publish(uuid, data)`
fans out to all subscribers registered for that UUID, dropping data
non-blockingly if a subscriber's buffer is full. `stream.subscribe(uuid)`
returns a channel and an idempotent `unsubscribe` function.

The four global channels in `apiHandler`, `apiHandlerv2`, `commonHandler`,
and `adminHandler` are replaced with `*stream` values. Admin SSE handlers
(`deviceLogsGet`, `deviceInfoGet`, `deviceRequestsGet`, `deviceMetricsGet`)
now subscribe to the stream keyed by the requested device UUID, so each
admin connection observes only its target device's data.

A `GET /admin/device/{uuid}/metrics` endpoint is added, completing the
per-device admin streaming surface alongside the existing logs, info, and
requests endpoints.

The `recordClient()` path in `apiHandler` and `apiHandlerv2` now publishes
request events to the stream (previously these were written to storage but
never forwarded to the admin channel).

## Thread-safety for all three storage drivers

All three `DeviceManager` implementations were unprotected against
concurrent map access. Under load from multiple simultaneous EVE devices
this caused data races, map-iteration panics, and corrupted state.

A plain `sync.Mutex` (not `RWMutex`) is used in each driver. EVE devices
are dominant writers -- the admin interface performs reads only occasionally
-- so the added complexity of a read-write lock would not be justified.

**`pkg/driver/file/device_manager_file.go`**

- Added `m sync.Mutex` protecting the three in-memory maps
  (`onboardCerts`, `deviceCerts`, `devices`).
- `refreshCache` gains an `isLocked bool` parameter. When `true` the
  caller already holds `d.m`, allowing a single lock span across both the
  cache refresh and subsequent map access (avoiding a redundant
  lock/unlock cycle). When `false` the function locks internally.
- Hot-path `Write*` methods lock only long enough to snapshot the
  `DeviceStorage` pointer from the map, then release before filesystem I/O.
- Internal helpers (`deviceExists`, `appExists`, `checkValidOnboardSerial`,
  `getOnboardSerialDevice`, `initDevice`) are lock-free and carry
  `// Caller must hold d.m.` comments.

**`pkg/driver/redis/device_manager_redis.go`**

- Same `m sync.Mutex` + `refreshCache(isLocked bool)` pattern as the
  file driver.
- Hot-path `Write*` methods lock briefly to get the Redis stream key,
  then release before network I/O.
- Fixed a latent nil-pointer dereference in `getOnboardSerialDevice` when
  a device record has no onboard certificate.

**`pkg/driver/memory/device_manager_memory.go`** and
**`pkg/driver/memory/byteslice.go`**

- Same `m sync.Mutex` on `DeviceManager`.
- Methods that previously delegated to other public methods
  (`OnboardRemove`, `DeviceRemove`, `DeviceRegister`, `GetDeviceOptions`,
  `GetGlobalOptions`) have those calls inlined to prevent self-deadlock.
- Added `sync.Mutex` to `ByteSlice` (`Write` and `Next` methods). Unlike
  the file driver's `ManagedFile`, which relies on OS-level file
  atomicity, `ByteSlice` is a pure in-memory structure and requires its
  own lock when a writer goroutine and a reader goroutine access the same
  instance concurrently.

## Concurrency tests and CI

Concurrent-access test suites added for all three drivers
(`TestDeviceManagerFileConcurrency`, `TestDeviceManagerRedisConcurrency`,
`TestDeviceManagerMemoryConcurrency`). Each suite covers:

- Concurrent writes to different devices
- Concurrent write types on the same device
- Device clear racing with active writes
- Device list racing with device clear
- Onboard register/remove/clear running concurrently
- Concurrent writers and readers for the same device

The `test` target in `Makefile` now passes `-race` to catch data races in
CI (`go test -race ./...`).


pkg/driver/file/device_manager_file.go
```diff
@@ -137,25 +138,20 @@ func (m *ManagedFile) Reader() (common.ChunkReader, error) {
...
-	maxFlowMessageSize      int
-	maxAppLogsSize          int
-	currentLogFile          *os.File
-	currentInfoFile         *os.File
-	currentMetricFile       *os.File
-	currentRequestsFile     *os.File
-	currentLogFileSize      int
-	currentInfoFileSize     int
-	currentMetricFileSize   int
-	currentRequestsFileSize int
```
milan-zededa, line 158: Here I also removed many unused struct fields

milan-zededa:
Still working on this, ignore this PR for now.

pkg/driver/file/device_manager_file.go
```diff
@@ -585,11 +584,6 @@ func (d *DeviceManager) initDevice(u uuid.UUID) error {
 
 // DeviceRegister register a new device cert
 func (d *DeviceManager) DeviceRegister(unew uuid.UUID, cert, onboard *x509.Certificate, serial string, conf []byte) error {
-	// refresh certs from filesystem, if needed - includes checking if necessary based on timer
-	err := d.refreshCache()
```
milan-zededa, line 589: This was redundant, DeviceRegister does not depend on the cache content

pkg/driver/file/device_manager_file.go
```diff
@@ -934,6 +950,9 @@ func (d *DeviceManager) refreshCache() error {
 		}
 		// convert into a certificate
 		certPem, _ := pem.Decode(b)
+		if certPem == nil {
+			return fmt.Errorf("unable to decode onboard certificate PEM data from file %s: %v", f, err)
```
milan-zededa, line 954: This error check was missing and there was a possibility of nil-pointer dereference

pkg/driver/file/device_manager_file.go
```diff
@@ -379,7 +374,9 @@ func (d *DeviceManager) OnboardClear() error {
 			return fmt.Errorf("unable to remove the onboard directory: %v", err)
 		}
 	}
+	d.m.Lock()
 	d.onboardCerts = map[string]map[string]bool{}
```
uncleDecart, line 378: why not sync.Map or RWMutex?

pkg/driver/file/device_manager_file.go
```diff
@@ -759,6 +767,7 @@ func (d *DeviceManager) WriteAppInstanceLogs(instanceID uuid.UUID, deviceID uuid
 		}
 	}
 	dev := d.devices[deviceID]
+	d.m.Unlock()
```
uncleDecart, line 764: why explicitly unclocking and not deferring? Shouldn't cause double locking on dev.AddAppLog or am I missing something?

pkg/driver/file/device_manager_file.go
```diff
@@ -1166,29 +1188,50 @@ func exists(path string) (bool, error) {
...
-	return d.devices[u].Info.Reader()
+	return dev.Info.Reader()
+}
+
+// GetMetricsReader returns a metrics reader for the specified device.
+func (d *DeviceManager) GetMetricsReader(u uuid.UUID) (common.ChunkReader, error) {
+	d.m.Lock()
+	exists := d.deviceExists(u)
+	dev := d.devices[u]
+	d.m.Unlock()
```
uncleDecart: Would it make sense to get it into a short function? I see this pattern repeats multiple times

uncleDecart (dismissed):
LGTM

pkg/driver/file/device_manager_file.go
milan-zededa: Using sync.Map for each field would protect individual map access, but it would not guarantee consistency across all internal fields of DeviceManager (onboardCerts, deviceCerts, and the devices map). For example, refreshCache rebuilds all three maps, and this needs to happen atomically, which is why a mutex is used.

As for the mutex type, RWMutex would not significantly improve parallelism. From the DeviceManager’s perspective, EVE devices (potentially many) act as writers, while the test framework (typically a single client of the Adam's /manage API) is the reader. This is the opposite of the scenario where an RW mutex is most beneficial.

More broadly, I prefer to keep this simple, even at the expense of parallelism. Adam is not designed for scale; it is intended for testing small deployments, typically a single EVE node or a cluster of 3 nodes.

pkg/driver/file/device_manager_file.go
milan-zededa: Keeping the lock until the end of the function wouldn’t break anything. I just aimed to limit the mutex to protecting internal state, which is what it’s intended for, rather than extending it over file I/O operations where it isn’t needed.

pkg/driver/file/device_manager_file.go
uncleDecart: point taken regarding consistency, without that sync.Map would be better choice (multiple writers single reader)

pkg/driver/file/device_manager_file.go
uncleDecart: okay, this is nitpicking and question of taste, let's leave it as it is

pkg/driver/file/device_manager_file.go
milan-zededa: Makes sense, done.

milan-zededa: @uncleDecart When you get a chance, please re-approve. Your previous approval got nullified after I pushed these changes.

uncleDecart (approved):

PR #148, opened 2026-04-19
Support multiple listen IP addresses

milan-zededa:
Allow Adam to bind to multiple IP addresses simultaneously by accepting
repeated `--ip` flags (`--ip 192.0.2.1 --ip ::1`) or a comma-separated list
(`--ip 192.0.2.1,::1`). The primary motivation is dual-stack testing:
binding to both an IPv4 and an IPv6 address lets developers verify
EVE-to-controller connectivity over both IP versions without running
separate Adam instances.
    
The change is backward compatible: existing deployments that pass a
single `--ip` value continue to work without modification.

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -10,6 +10,7 @@ import (
 	"log"
 	"os"
 	"path"
+	"strings"
 
 	"github.com/lf-edge/adam/pkg/driver/common"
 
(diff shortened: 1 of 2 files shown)


uncleDecart (approved):
LGTM, since we're not expecting hundreds of connections there's no need to restrict number of IP addresses

PR #150, opened 2026-04-28
Do not create base device config

milan-zededa:
Previously, Adam automatically created a minimal "base" config for every newly registered device. This caused two problems:

1. It made it impossible to test the scenario where the controller has no config for a device yet, which is a legitimate scenario that test frameworks need to exercise.

2. Devices can be provisioned with a bootstrap config baked in, which is necessary when the default network setup (DHCP on all Ethernet interfaces) does not provide controller connectivity -- for example when the network uses static IP addresses, VLANs, or a HTTP proxy for management traffic. If the test framework was too slow to submit the initial config, the device would fetch the auto-generated base config first, effectively overwriting the bootstrap config (and breaking connectivity for a while, until network config fallback kicks in). This forced the test framework to race against the first device config poll, an inherently fragile arrangement.

With this change, `GetConfig` returns `NotFoundError` when no config has been submitted yet. The API handlers propagate this as an HTTP 404, allowing the device to retry. `GetDeviceOptions` and `GetGlobalOptions` likewise return `NotFoundError` instead of auto-creating defaults; their callers treat not-found as "use zero-value defaults".

I have also enhanced the CI workflow to include running a Redis container. Previously, all Redis tests were silently skipped because no Redis instance was available.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,6 +10,17 @@ on:
 jobs:
   build:
     runs-on: ubuntu-24.04
+    services:
+      # Redis is required so that the Redis driver unit tests are not skipped.
+      redis:
+        image: redis:7
+        ports:
+          - 6379:6379
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 5s
+          --health-timeout 3s
+          --health-retries 5
     steps:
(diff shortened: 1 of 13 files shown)


eriknordmark (approved):
LGTM

PR #151, opened 2026-04-28
Add admin API endpoints for app logs and flow records

milan-zededa:
The [eden](https://github.com/lf-edge/eden) testing framework currently retrieves app logs and flow records by directly reading Adam's on-disk database. This couples eden to Adam's internal storage layout and makes it fragile to any change in how Adam stores data. This PR adds proper admin API endpoints for both data types so that external tools can fetch them through Adam's public interface.

- Add `GET /admin/device/{uuid}/app/{appuuid}/logs`: returns accumulated app instance logs or streams them in real time (via `Stream: true` header), consistent with the existing device-level log/info/metric endpoints.
- Add `GET /admin/device/{uuid}/flowlogs`: same one-shot/streaming semantics for flow records.
- Extend the `DeviceManager` interface with `GetAppLogsReader` and `GetFlowMessageReader`; implement both in the File, Memory, and Redis drivers.
- Generalize the in-memory pub/sub stream to key subscribers by an `instanceID` (device UUID + optional app UUID), enabling per-app streaming without a separate stream object per endpoint.
- Unify storage format: info and metrics are now converted to JSON at write time (matching logs and app logs), removing the read-time `conversionFunc` callback from `deviceDataGet` and fixing a silent data-loss bug in the file driver where concatenated protobuf messages were not self-delimiting.
- Fix two file driver bugs where required subdirectories (`app/<uuid>/logs`, `flow_message`) were never created on disk, causing the first write to fail with "no such file or directory".
- Fix `openTimestampFile` which used `.111` as a Go time-format literal (not milliseconds), causing files created within the same second to overwrite each other.

### DIFF
--- a/pkg/driver/common/common.go
+++ b/pkg/driver/common/common.go
@@ -44,6 +44,15 @@ type ChunkReader interface {
 	Next() (io.Reader, int64, error)
 }
 
+// EmptyChunkReader is a ChunkReader implementation that contains no chunks.
+// Its Next method always returns io.EOF, indicating that no data is available.
+type EmptyChunkReader struct{}
+
+// Next always returns io.EOF because there are no chunks.
+func (EmptyChunkReader) Next() (io.Reader, int64, error) {
+	return nil, 0, io.EOF
+}
+
 type BigData interface {
 	Get(index int) ([]byte, error)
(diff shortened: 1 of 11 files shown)


eriknordmark (approved):
LGTM

